PR #12, opened 2017-11-14
Update README

drdavella:


README.rst
```diff
@@ -1,7 +1,210 @@
...
+value of ``__doctest_skip__`` should be a list of wildcard patterns for all
+functions/classes whose doctests should be skipped.  For example::
+
+   __doctest_skip__ = ['myfunction', 'MyClass', 'MyClass.*']
+
+skips the doctests in a function called ``myfunction``, the doctest for a
+class called ``MyClass``, and all *methods* of ``MyClass``.
+
+Module docstrings may contain doctests as well.  To skip the module-level
+doctests include the string ``'.'`` in ``__doctest_skip__``.
```
pllim: Does this mean `__doctest_skip__ = ['.']`? If so, it is less confusing to include this as explicit example like the others.

README.rst
```diff
@@ -1,7 +1,210 @@
...
+
+To skip all doctests in a module::
+
+   __doctest_skip__ = ['*']
+
+Doctest Dependencies
```
pllim: I don't see any mention of `.. doctest-requires:: scipy` in this section but it is used in `astropy`. Is that still supported?

README.rst
```diff
@@ -1,7 +1,210 @@
...
+floating point values.  Floating point representations are often not exact and
+contain roundoffs in their least significant digits.  Depending on the platform
+the tests are being run on (different Python versions, different OS, etc.) the
+exact number of digits shown can differ.  Because doctests work by comparing
+strings this can cause such tests to fail.
+
+To address this issue, the ``pytest-doctestplus`` plugin provides support for a
+``FLOAT_CMP`` flag that can be used with doctests.  For example:
+
+.. code-block:: none
```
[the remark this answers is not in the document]
drdavella: Will fix. This was actually copied from astropy's documentation though.

README.rst
drdavella: Good point. They should be included here.

README.rst
```diff
@@ -1,7 +1,210 @@
...
+
+This plugin provides advanced features for testing example Python code that is
+included in Python docstrings and in standalone documentation files.
+
+Good documentation for developers contains example code. This is true of both
+standalone documentation and of documentation that is integrated with the code
+itself. Python provides a mechanism for testing code snippets that are provided
+in Python docstrings. Pytest provides a mechanism for running doctests against
+both docstrings in source code and in standalone documentation files. These are
+Good Things.
```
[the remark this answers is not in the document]
drdavella: Probably not, but sometimes it seems like such things need to be said... 😉

README.rst
drdavella: Actually, I'm not sure we should document these here quite yet since I think they also require astropy's sphinx plugins in order to work, and we haven't provided separate repositories for those yet.

drdavella: I opened an issue to track this in #13.

PR #23, opened 2018-08-15
Include tests in sdists

toddrme2178:
Tests are used by some downstreams to verify the build was done correctly.


drdavella:
This seems perfectly reasonable to me as long as there is no expectation for the tests to actually be installed.

PR #34, opened 2019-01-03
Honor the collect_ignore option used in conftest.py

drdavella:
This resolves #33.

@oscarbenjamin, if you have a chance would you be able to test this in your repo (without using `norecursedirs`)?

PR #36, opened 2019-01-03
Honor the collect_ignore option used in conftest.py

oscarbenjamin:
Replacement for #34

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.3.0 (unreleased)
 ==================
 
+- Honor the ``collect_ignore`` option used in ``conftest.py``. [#36]
+
 0.2.0 (2018-11-14)
 ==================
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,8 @@
 0.3.0 (unreleased)
 ==================
 
+- Honor the ``collect_ignore`` option used in ``conftest.py``. [#34]
```
drdavella: Just a nit, but we need to change the PR number here now.

oscarbenjamin: Okay, I've just fixed that and force-pushed.

oscarbenjamin:
I've pushed another change because it would have failed the tests - if `collect_ignore` is not specified in `conftest.py` then `_getcontest_pathlist` returns `None` rather than an empty list.

drdavella:
Thank you @oscarbenjamin!

PR #37, opened 2019-01-03
Inline np.allclose to remove dependency on numpy

oscarbenjamin:
Using this with sympy pulled in numpy as a dependency. It seems unnecessary to have such a big dependency just for `allclose`. This patch eliminates that dependency by reimplementing allclose for scalars.

The implementation is based on the [documented formula for allclose](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.allclose.html)

### DIFF
--- a/pytest_doctestplus/output_checker.py
+++ b/pytest_doctestplus/output_checker.py
@@ -6,8 +6,7 @@ normalizations of Python expression output.  See the docstring on
 
 import doctest
 import re
-
-import numpy as np
+import math
 
 import six
 from six.moves import zip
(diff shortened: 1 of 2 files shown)


drdavella:
I think I'm okay with this. Any thoughts @astrofrog, @bsipocz?

oscarbenjamin:
Looks as if the test failed for an unrelated reason...

drdavella:
Yes, the test failure is unrelated. I merged a PR with a fix recently so if you rebased it should fix it, but it's not necessary.

PR #39, opened 2019-01-03
Make use of doctest_optionflags settings

### LINKED ISSUE #38: doctestplus ignores doctest_optionflags options in pytest.ini
I've almost got doctestplus and sympy to play nicely together but I have a few outstanding issues.

One is that I need a way to set the `IGNORE_EXCEPTION_DETAIL` doctest option. This is set in `pytest.ini` like
```ini
doctest_optionflags =
    NORMALIZE_WHITESPACE
    IGNORE_EXCEPTION_DETAIL
    ELLIPSIS
```

However doctestplus ignores these flags and always uses `NORMALIZEWHITESPACE` and `ELLIPSIS` but not `IGNORE_EXCEPTION_DETAIL`.

Would it not be better to get these flags from the pytest config?

I also want a way to set `FLOAT_CMP` globally - maybe it could go in `pytest.ini` with the other flags as well...

drdavella:
This resolves #38. @oscarbenjamin please take a look.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 
 - Honor the ``collect_ignore`` option used in ``conftest.py``. [#36]
 
+- Make use of ``doctest_optionflags`` settings. [#39]
+
 0.2.0 (2018-11-14)
 ==================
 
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
I've tested this with sympy comparing with master with and without the `IGNORE_EXCEPTION_DETAIL` setting in `pytest.ini`. It seems to work fine here.

Thanks for this!

PR #41, opened 2019-01-05
Add doctest_skip and doctest_skipif decorators

oscarbenjamin:
This patch adds two new decorators `doctest_skip` and `doctest_skipif` that can be used to mark an object so that doctests are not collected from it.

The `doctest_skip` decorator is for unconditional skipping so is partly redundant given `__doctest_skip__`. However the `doctest_skipif` decoartor is not redundant since doctestplus currently doesn't provide any way to skip based on arbitrary conditions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -189,6 +189,45 @@ Having this module-level variable will require ``scipy`` to be importable
 in order to run the doctests for functions ``func1`` and ``func2`` in that
 module.
 
+Doctest skip decorators
+^^^^^^^^^^^^^^^^^^^^^^^
+
+A function or class can be marked with the decorator ``@doctest_skip`` or
+``@doctest_skipif`` to inddicate that it should be skipped when collecting
+doctests.
+
+This can be used to unconditionally skip collection optionally providing a
+reason::
+
+    from pytest_doctestplus import doctest_skip
+
+    @doctest_skip(reason="Integer division")
+    def func1():
+        '''
+        >>> 1/2
+        0.5
+        '''
+        pass
+
+This can also be used to conditionally skip doctests::
+
+    import sys
+    from pytest_doctestplus import doctest_skipif
+
+    @doctest_skipif(sys.version_info < (3,0), "Integer division on Python 2.x")
+    def func1():
+        '''
+        >>> 1/2
+        0.5
+        '''
+        pass
+
+The ``doctest_skipif`` decorator can also accept a callable::
+
+    @doctest_skipif(lambda: not thing_installed(), "Need thing")
+    def func1():
+        ...
+
 Remote Data
 ~~~~~~~~~~~
 
--- a/pytest_doctestplus/__init__.py
+++ b/pytest_doctestplus/__init__.py
@@ -2,3 +2,5 @@
 """
 This package contains pytest plugins that are used by the astropy test suite.
 """
+
+from pytest_doctestplus.skipping import doctest_skip, doctest_skipif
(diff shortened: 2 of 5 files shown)


drdavella:
@oscarbenjamin I am fully in support of the idea behind this. However, I'm a little concerned that it's not terribly flexible. It can only be applied to functions and class definitions, which severely limits its applicability. I think it would be preferable to use a custom directive to handle this kind of logic and to control the execution of entire code blocks.

It seems possible to me to implement a directive like ``.. doctest-skipif`` that takes an argument, which could be an arbitrary python statement. We could `eval` this statement in the plugin to determine whether the code block should be skipped or not.

oscarbenjamin:
What scope would the condition be `eval`led in? The code that currently handles e.g. `.. doctest-skip` doesn't have access to any scope. Is the idea that you could do something like:

```python
def f():
    '''
    >>> from mod import check

    .. doctest-skipif:: check()
    >>> f()
    1
    '''
    return 1
```

oscarbenjamin:
If so then it would require reimplementing some of `DocTestRunner` because I think the condition would have to be `eval`ed in between iterations of this loop: [https://github.com/python/cpython/blob/master/Lib/doctest.py#L1293](https://github.com/python/cpython/blob/master/Lib/doctest.py#L1293)

drdavella:
@oscarbenjamin that's a good point. However, if I understand correctly, the plugin should have access to everything that's in scope for the module being tested (due to the way that `pytest` collection works). So I wonder if a simple `eval` in our plugin will work for most cases after all. But I might be wrong about that.

Just out of curiosity, is there any possibility that `pytest.mark.skip` and `pytest.mark.skipif` already work for the same cases where you've implemented `doctest_skip[if]`?

oscarbenjamin:
> @oscarbenjamin that's a good point. However, if I understand correctly, the plugin should have access to everything that's in scope for the module being tested (due to the way that `pytest` collection works). So I wonder if a simple `eval` in our plugin will work for most cases after all. But I might be wrong about that.

You have to provide the `globals` dict to `eval`: [https://docs.python.org/3/library/functions.html#eval](https://docs.python.org/3/library/functions.html#eval) to get it to execute in another scope.

> Just out of curiosity, is there any possibility that `pytest.mark.skip` and `pytest.mark.skipif` already work for the same cases where you've implemented `doctest_skip[if]`?

They don't work. Also `pytest.mark.skipif` suffers from the same problem that the condition has to be evaluated at import time. That's not a problem in `test_*.py` files because you don't import them during normal usage. Doctests are in the normal run-time files.

The directives '.. doctest-skip' etc only work in .rst files. The .rst files don't have any natural scope to evalute the condition in, especially at parse time. I am looking for something that can work in normal modules. There you would have a scope but the globals for that scope aren't available during parsing.

What I am particularly looking for is to be able to skip all doctests in a docstring conditionally like doctest-skipall-if. I think that might be easier to implement. Otherwise I think there needs to be a DocTestRunner class.

drdavella:
@oscarbenjamin sorry for getting back to this just now. I'm okay with moving ahead with this even though it's not quite as flexible as I had hoped. Given the constraints, it seems like the best we can do for now without significantly more effort.

oscarbenjamin:
Hi @drdavella, no problem. Actually I think it's maybe best just to leave this for now. In SymPy I've used the `__doctest_requires__` marker for this already now and I've got to the point where the tests are working - there are just presumably some corner-cases where some tests will fail because of missing dependencies in different setups. I'd rather wait to discuss what would be best with others in SymPy before pushing ahead with this PR.

drdavella:
@oscarbenjamin thanks for the update. In that case, maybe we should just close this for now? Feel free to reopen if/when you'd like to move forward with it.

oscarbenjamin:
Closing for now

PR #65, opened 2019-08-28
TST: Move Windows testing to Travis, remove Appveyor

pllim:
I don't know what the `platform` thingy does in `Appveyor.yml`, so I just ignored it. Over all, Appveyor matrix looks similar to the Travis one, so my intent here is to run the exact same set of matrix twice, once on Linux and once on Windows. But if Windows jobs on Travis turn out to be too slow, we can either:

* Scale down what to test on Windows in Travis, **or**
* Port everything to Azure.


astrofrog:
We could probably make things a lot faster by not using conda... :trollface:

PR #82, opened 2019-11-07
respect ignore and ignore-glob options

### LINKED ISSUE #58: Respect `ignore`
https://github.com/coala/coala-bears setup.cfg includes `--ignore=tests/python/bandit_test_files/` , as they are unparsable python code.  When I try to use doctestplus, these files are found and imported, resulting in an error.

vanyakosmos:
close #58

### DIFF
--- a/pytest_doctestplus/plugin.py
+++ b/pytest_doctestplus/plugin.py
@@ -3,14 +3,13 @@
 This plugin provides advanced doctest support and enables the testing of .rst
 files.
 """
-import six
-
 import doctest
 import fnmatch
 import os
 import re
 import sys
 import warnings
+from distutils.version import LooseVersion
 
 try:
     from textwrap import indent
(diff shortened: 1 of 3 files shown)


pytest_doctestplus/plugin.py
```diff
@@ -49,13 +49,22 @@ def pytest_addoption(parser):
...
 
+    # add ignore-glob option for old versions of pytest
+    if LooseVersion('4.3.0') > LooseVersion(pytest.__version__):
+        parser.addoption(
+            "--ignore-glob",
+            action="append",
+            metavar="path",
+            help="ignore path pattern during collection (multi-allowed).",
+        )
+
```
astrofrog: I wonder if it'd be better to just not support this option on older versions of pytest since otherwise we'll get inconsistent behavior where this works for doctests but not for regular ``test_*.py`` files?

astrofrog:
@vanyakosmos - could you rebase this PR? Thanks!

pytest_doctestplus/plugin.py
```diff
@@ -88,11 +89,13 @@ def pytest_addoption(parser):
                   "Default format for docs. "
                   "This is no longer recommended, use --doctest-glob instead.")
 
+    parser.addini("text_file_format", "changing default format for docs")
```
astrofrog: Remove this? (it looks like it is already defined above)

vanyakosmos: oh, i probably messed up rebase

PR #113, opened 2020-05-20
Add the doctestplus sphinx extension

Cadair:
I could see no good reason why this extension was in `sphinx-astropy` rather than `pytest_doctestplus` if all it is doing is telling sphinx to ignore the doctestplus metadata?

I made one small modification to this, which is to set the `override=True` flag on the two directives which are also defined in the sphinx doctest extension, this prevents any warnings being raised when parsing the conf.py file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -250,6 +250,17 @@ statements marked with ``--remote-data`` will be skipped.
 .. _pytest-remotedata: https://github.com/astropy/pytest-remotedata
 __ pytest-remotedata_
 
+Sphinx Compatibility
+~~~~~~~~~~~~~~~~~~~~
+
+To use the additional directives when building your documentation with sphinx
+you may want to enable the sphinx extension which registers these directives
+with sphinx. Doing so ensures that sphinx correctly ignores these directives,
+running the doctests with sphinx is not supported. To do this, add
+``'pytest_doctestplus.sphinx.doctestplus'`` to your ``extensions`` list in your
+``conf.py`` file.
+
+
 Development Status
 ------------------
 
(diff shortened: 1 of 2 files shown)


pllim:
Not sure who should review this. @astrofrog ?

pllim:
Do you have another PR to take it out of sphinx-astropy?

pllim:
What if someone uses sphinx-astropy and this plugin and both of them have the same thing?

Cadair:
> What if someone uses sphinx-astropy and this plugin and both of them have the same thing?

You need to enable this extension in your conf.py, so it wont just break if it's installed. Once this has made it onto pypi I can open a PR to sphinx-astropy to swap it over.

astrofrog:
I think it's a little strange to have a package called pytest-doctestplus be a sphinx extension though. Should we consider renaming this package to simply 'doctestplus'? At the end of the day I can live with this here and not change the package name but just want to express that something doesn't feel quite right ;)

pllim:
Re: renaming package -- But it will never run with other test runners, right? I don't see the point to make things complicated for everyone, so I say we keep the name. But regardless, it should be discussed as a new issue.

Cadair:
Well the sphinx extension doesn't actually *do* anything, it just stops sphinx complaining that the directives ingested by pytest-doctestplus don't exist.

astrofrog:
Ok I can live with the name :)

Cadair:
but @astrofrog can you live with the PR?

astrofrog (commented):
Is there any way to test this? 😁

Cadair:
Not by calling sphinx that I can think of...

astrofrog (approved):

astrofrog:
Will you do a matching PR in Sphinx-astropy?

Cadair:
Once this is on pypi I can update sphinx-astropy

saimn:
Can we merge this ?

Cadair:
yes.

saimn:
Needs a changelog entry then ;)

PR #136, opened 2020-12-11
Add flag to show warnings

saimn:
For https://github.com/astropy/astropy/pull/11137
Basically copy/pasted from IGNORE_WARNINGS

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -151,8 +151,10 @@ Showing warnings
 ~~~~~~~~~~~~~~~~
 
 If code in a doctest emits a warning and you want to make sure that warning is
-shown, even if warnings are turned into errors by pytest, you can make use of
-the ``SHOW_WARNINGS`` flag. For example:
+shown, you can make use of the ``SHOW_WARNINGS`` flag. This is useful when
+warnings are turned into errors by pytest, and also because by default warnings
+are printed to stderr. This is the opposite from ``IGNORE_WARNINGS`` so
+obviously the two flags should not be used together. For example:
 
 .. code-block:: python
 
--- a/README.rst
+++ b/README.rst
@@ -147,6 +147,21 @@ you can make use of the ``IGNORE_WARNINGS`` flag. For example:
   >>> np.mean([])  # doctest: +IGNORE_WARNINGS
   np.nan
 
+Showing warnings
+~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


pllim (commented):
Not clear from the doc if it can be used together with `IGNORE_WARNINGS` or is mutually exclusive with it.

Also would be curious to know how it interacts with `filter_warnings` set in `setup.cfg`, or inside a `warnings.catch_warnings` context manager.

saimn:
> Not clear from the doc if it can be used together with IGNORE_WARNINGS or is mutually exclusive with it.

It doesn't make sense to use both. But if you do I guess the inner call will win.

> Also would be curious to know how it interacts with filter_warnings set in setup.cfg, 

The goal is actually to override the default settings in setup.cfg (which for astropy turns warnings into exceptions)

> or inside a warnings.catch_warnings context manager.

It wraps the code, so the warnings that would be catch manually with catch_warnings would not be seen by the doctest flag.

pllim:
Re: https://github.com/astropy/pytest-doctestplus/pull/136#issuecomment-743247354 -- Can we document this in README? 😸

README.rst
```diff
@@ -147,6 +147,23 @@ you can make use of the ``IGNORE_WARNINGS`` flag. For example:
...
+If code in a doctest emits a warning and you want to make sure that warning is
+shown, you can make use of the ``SHOW_WARNINGS`` flag. This is useful when
+warnings are turned into errors by pytest, and also because by default warnings
+are printed to stderr. This is the opposite from ``IGNORE_WARNINGS`` so
+obviously the two flags should not be used together. For example:
+
+.. code-block:: python
+
+  >>> import numpy as np
+  >>> np.mean([])  # doctest: +SHOW_WARNINGS
```
pllim, line 162: One more question: If the expected warning does not show, will doctest fail or ignore?

Is this the doctest equivalent of `pytest.warns`?

saimn: > One more question: If the expected warning does not show, will doctest fail or ignore?

Yep. I added more tests.

> Is this the doctest equivalent of pytest.warns?

Yes, it can be seen as such.

pllim (approved):
Thanks! LGTM. I'll leave it open in case @astrofrog wants to review too.

astrofrog (approved):
Looks good to me, thanks!

saimn:
Thanks. Will need a release ;)

PR #139, opened 2020-12-15
Add workflow to publish on PyPI

saimn:


saimn:
Tested on Test PyPI with my fork : https://github.com/saimn/pytest-doctestplus/runs/1553716657?check_suite_focus=true
(I deleted the project on Test PyPI afterwards, which was not a great idea since the name cannot be reused now :()

pllim:
> the name cannot be reused now

Anyway we can contact PyPI to reclaim the name or something? Would be nice to have it back for future testing...

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
```
pllim, line 1: Maybe something shorter and does not include emoji (shocking coming from me, I know)? This name is the same name used in badge, in case we decide to include this in a badge in the future.
```suggestion
name: Release
```

saimn:
It seems not: https://test.pypi.org/help/#file-name-reuse (but we could still use another project, or another name, so even if it's a bit annoying)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  push:
+    tags:
+      - '*'
```
pllim, line 6: Do we ever need to push a tag without releasing?

Do we need to disable this in forks? Can we?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
```
pllim, line 10:
```suggestion
    name: Build and publish Python 🐍 distributions 📦 to PyPI
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+    - uses: actions/checkout@master
+      with:
+        fetch-depth: 0
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Install python-build and twine
+      run: python -m pip install build twine
```
pllim, line 22: Do we need to add `wheel` too or is that somehow automatically included?
```suggestion
      run: python -m pip install build twine wheel
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Build package
+      run: python -m build --sdist --wheel .
+
+    - name: List result
+      run: ls -l dist
+
+    - name: Check long_description
+      run: python -m twine check dist/*
```
pllim, line 31: If you want the publish to not happen if there are warnings, it has to be strict.
```suggestion
      run: python -m twine check dist/* --strict
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Test package
+      run: |
+        cd ..
+        python -m venv testenv
+        testenv/bin/pip install pytest pytest-doctestplus/dist/*.whl
+        testenv/bin/pytest pytest-doctestplus/tests --doctest-plus --doctest-rst
+
+    - name: Publish distribution 📦 to PyPI
+      uses: pypa/gh-action-pypi-publish@master
```
pllim, line 41: Is it safer to use a "stable" tag?
```suggestion
      uses: pypa/gh-action-pypi-publish@v1
```

pllim (commented):
Thanks! I like the concept and my comments are mostly minor.

CI failure is real but unrelated: `AttributeError: 'LocalPath' object has no attribute 'is_file'`

But that failure also blocks the actual release, as in we should not release until it is fixed. (See #140)

p.s. Unfortunate about not able to claim the test namespace back but good to know we have a workaround.

saimn:
> CI failure is real but unrelated: AttributeError: 'LocalPath' object has no attribute 'is_file'

Yes, seems related to some internal changes (https://github.com/pytest-dev/pytest/pull/8122), hopefully this will disappear...

.github/workflows/publish.yml
saimn: > Do we ever need to push a tag without releasing?

I don't understand ?

> Do we need to disable this in forks? Can we?

Forks will not have the secret to upload, and should not push tags on their repo, so  think it's fine.

.github/workflows/publish.yml
saimn: We don't need it, the sdist and wheel are produced with "build".

.github/workflows/publish.yml
saimn: Currently there is warning about "long description type", not sure if we want to fail for this ? Oh, and I don't see `--strict` in `twine check`'s help ?

.github/workflows/publish.yml
pllim: > I don't understand ?

Will it ever be a situation where one would do something like `git tag aas-tutorial` or something crazy, where we need to push a tag but it is not for release? If not, then it's fine.

pllim:
> hopefully this will disappear

I am not sure if it will disappear if the change was intentional. Should we file a bug report with `pytest`?

.github/workflows/publish.yml
saimn: Yeah probably.

.github/workflows/publish.yml
saimn: Ah ok, I don't think in a general case. And the build would probably fail, since the package version that would be generated by setuptools_scm is by default not a valid version accepted by PyPI (when it contains the commit hash for dev versions).

saimn:
> I am not sure if it will disappear if the change was intentional. Should we file a bug report with pytest?

It's still failing today after the commit with your suggestions, so maybe good to report it indeed.

.github/workflows/publish.yml
saimn: So to make sure the workflow works, I'm thinking that we could run it on push events (not sure if that include merges on master ?), and just skip the upload part if the push is not a tag with:
```
  if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
Thoughts?

pllim: Can we just test it once in this PR before merge by adding push/pull request (not sure which one) but comment out the upload part? Is there a reason why the testing needs to be done for every push to master?

Yes, I think "push" includes merge commits.

.github/workflows/publish.yml
pllim: Hmm I see it here: https://twine.readthedocs.io/en/latest/#twine-check

.github/workflows/publish.yml
saimn: > Is there a reason why the testing needs to be done for every push to master?

Just to be sure it continues to work in the future, and avoid finding issues when pushing a tag. Running it for pushes/merges on master seems a good solution ? (Or it could also be another build for pull requests)

pllim: How about "push tag only + scheduled cron" instead of blind "push"? And we skip upload on cron.

saimn: That could work as well. I would prefer push because it would run just after a PR is merged and also if you push directly a README/Changelog update just before a release.

.github/workflows/publish.yml
saimn: Oh, funny:
```
❯ twine check --help
usage: twine check [-h] dist [dist ...]

positional arguments:
  dist        The distribution files to check, usually dist/*

optional arguments:
  -h, --help  show this help message and exit
```

pllim: Ah... Maybe pypa/twine#715 is not released yet. It is milestoned to 3.3.0 and current release is 3.2.0.

.github/workflows/publish.yml
pllim: Since you have a strong preference to using `push`, let's try it your way and we can revisit if needed. 😄

saimn: Ok ;), let's try this.

.github/workflows/publish.yml
saimn: Hmm, I run temporarily the workflow on pull request event, and get "unable to find version `v1`".

pllim: Hmm! Maybe they don't support it. Now that I look at their README, it does say `@master`. Sorry for the noise!

saimn:
I now get a `'PosixPath' object has no attribute 'isfile'`, and the same locally on master.

saimn:
Because 6.2.1 does not have the pathlib changes.

saimn:
Ok, this time it works, and skips the upload: https://github.com/astropy/pytest-doctestplus/pull/139/checks?check_run_id=1561218146
I will remove the pull request event.

pllim (approved):
Thanks! 🚀

saimn:
Cool ! And the job ran on master (https://github.com/astropy/pytest-doctestplus/runs/1564407053?check_suite_focus=true).
So can we test this to make a release ? ;)

PR #140, opened 2020-12-15
MNT: Compat with pytest>=6.3

pllim:
As recommended in https://github.com/pytest-dev/pytest/pull/8122#issuecomment-745450317


saimn (approved):
Looks good, given the discussions on the pytest ticket. Thanks @pllim !

PR #141, opened 2020-12-16
Fix compatibility with pytest dev/6.2

saimn:


pytest_doctestplus/plugin.py
```diff
@@ -22,7 +22,7 @@
 
 _pytest_version = Version(pytest.__version__)
 PYTEST_GT_5 = _pytest_version > Version('5.9.9')
-PYTEST_GT_6_2 = _pytest_version > Version('6.2')
+PYTEST_GE_6_3 = _pytest_version >= Version('6.3')
```
pllim, line 25: This was what I actually started with, but somehow didn't work.

saimn:
It gets more annoying ...
```
(Pdb) Version(pytest.__version__)
<Version('6.3.0.dev19+g950fbbc32')>
(Pdb) Version(pytest.__version__) >= Version('6.3')
False
```

pytest_doctestplus/plugin.py
saimn: Yep, I see that :(

pytest_doctestplus/plugin.py
```diff
@@ -22,7 +22,7 @@
 
 _pytest_version = Version(pytest.__version__)
 PYTEST_GT_5 = _pytest_version > Version('5.9.9')
-PYTEST_GE_6_3 = _pytest_version >= Version('6.3')
+PYTEST_GE_6_3 = _pytest_version.is_devrelease or _pytest_version >= Version('6.3')
```
pllim, line 25: Or something hacky like this if this doesn't work... 😬 
```python
PYTEST_GT_6_2 = _pytest_version >= Version('6.2.99')
```

saimn: It works ;), so maybe good enough for now ?

pllim (approved):
Thanks for cleaning up my mess!

saimn:
My pleasure ;)

PR #148, opened 2021-01-16
Fix issue with IGNORE_WARNINGS and SHOW_WARNINGS and skipped blocks

### LINKED ISSUE #147: Fix usage of `+IGNORE_WARNING` directive with the `..doctest-remote-data ::` directive
@saimn @pllim - we're kind of stuck here. This code chunck is skipped as expected with the new directive, but the test comes back as "passed" rather than "skipped". Have you recall any similar situation, or the reason for this?

_Originally posted by @bsipocz in https://github.com/astropy/pytest-doctestplus/pull/137#discussion_r551495732_

saimn:
Fix #147.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,9 @@
 0.10.0 (unreleased)
 ===================
 
+- Fix wrong behavior with ``IGNORE_WARNINGS`` and ``SHOW_WARNINGS`` that could
+  make a block to pass instead of being skipped. [#148]
+
 0.9.0 (2021-01-14)
 ==================
 
--- a/pytest_doctestplus/plugin.py
+++ b/pytest_doctestplus/plugin.py
@@ -372,25 +372,26 @@ def pytest_configure(config):
                         required = re.split(r'\s*[,\s]\s*', match.group(1))
                 elif isinstance(entry, doctest.Example):
 
+                    has_required_modules = DocTestFinderPlus.check_required_modules(required)
+                    if skip_all or skip_next or not has_required_modules:
+                        entry.options[doctest.SKIP] = True
+
+                    elif (config.getoption('remote_data', 'none') != 'any'
+                            and entry.options.get(REMOTE_DATA)):
+                        entry.options[doctest.SKIP] = True
+
                     # If warnings are to be ignored we need to catch them by
                     # wrapping the source in a context manager.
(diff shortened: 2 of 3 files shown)


pytest_doctestplus/plugin.py
```diff
@@ -372,26 +372,27 @@ def parse(self, s, name=None):
...
                                         + indent(entry.source, '    '))
                         show_warnings_context_needed = True
 
-                    has_required_modules = DocTestFinderPlus.check_required_modules(required)
-                    if skip_all or skip_next or not has_required_modules:
-                        entry.options[doctest.SKIP] = True
-
-                    if config.getoption('remote_data', 'none') != 'any' and entry.options.get(REMOTE_DATA):
-                        entry.options[doctest.SKIP] = True
-
```
saimn, line 394: So the problem was that IGNORE_WARNINGS (and SHOW_WARNINGS) were always checked, so even if the block was skipped a `doctest.Example` with the IGNORE_WARNINGS_CONTEXT was added (below) which would make the full block to be passed instead of skipped.

CHANGES.rst
```diff
@@ -1,6 +1,9 @@
 0.10.0 (unreleased)
 ===================
 
+- Fix wrong behavior with ``IGNORE_WARNINGS`` and ``SHOW_WARNINGS`` that could
+  make a block to pass instead of being skipped. [GUESSPR]
```
pllim, line 5: Did I win? 😹 
```suggestion
  make a block to pass instead of being skipped. [#148]
```

pytest_doctestplus/plugin.py
pllim: Hmm... What if someone wants to skip warning *inside* a remote data block (maybe out of scope here and to be handled in #137) or the "require" block?

CHANGES.rst
saimn: Oups, I don't have the pre-commit hook on this repo ;)

pytest_doctestplus/plugin.py
saimn: That should work, the remote data block is moved above, just to prevent adding the warnings code if the block is already marked as skipped. But if the block is not skipped then the warnings code will be added.

pllim (approved):
Yeah, okay, that makes sense. Merging then, so #137 can build on top of this. Thank you!

PR #152, opened 2021-02-17
Fixed logic for IGNORE_WARNINGS/SHOW_WARNINGS for --remote-data=any

ayshih:
This PR fixes a bug where the handling of `IGNORE_WARNINGS`/`SHOW_WARNINGS` flags in module docstrings is entirely skipped over if `--remote-data=any` is specified (for concurrent use of the `pytest-remotedata` plugin).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 0.10.0 (unreleased)
 ===================
 
+- Fixed a bug where the command-line option ``--remote-data=any`` (associated
+  with the ``pytest-remotedata`` plugin) would cause ``IGNORE_WARNINGS`` and
+  ``SHOW_WARNINGS`` options to be ignored in module docstrings. [#152]
+
 - Fix wrong behavior with ``IGNORE_WARNINGS`` and ``SHOW_WARNINGS`` that could
   make a block to pass instead of being skipped. [#148]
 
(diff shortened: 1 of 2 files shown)


pllim:
Huh, why does this sound familiar? Did #148 not work for you?

ayshih:
#148 fixes a loosely related bug for RST files, but this bug is for module docstrings.

saimn:
Oh indeed, I saw this `if` once and did not understand why it was like this. Good catch, and thanks for the PR @ayshih !

saimn (approved):

saimn:
While pushing the merge button I was thinking that it would be great if there was a test for this ;-). A bit late for this but if you have a bit more time to spare on this, would be appreciated.

ayshih:
I contemplated adding a test, but it's a little weird because I don't believe you can actually feed in `--remote-data=any` as a command-line option without also having `pytest-remotedata` (i.e., it errors).  Do you want that dependency in the test suite?  Or should the test instead inject the command-line option after `pytest` has already checked for validity of the arguments?

pllim:
I think a test can be added as part of #137 or as a follow up to it?

saimn:
Ah right, it's not there yet. Yes, could be in #137 then.

PR #156, opened 2021-03-09
TST: pytest-dev compat

### LINKED ISSUE #155: DeprecationWarning from pytest-dev
```
_________________ ERROR collecting test_show_warnings_rst.rst __________________
.../pytest_doctestplus/plugin.py:266: in collect
    text = self.fspath.read_text(encoding)
.../_pytest/nodes.py:202: in fspath
    warnings.warn(NODE_FSPATH.format(type=type(self).__name__), stacklevel=2)
E   pytest.PytestDeprecationWarning: DocTestTextfilePlus.fspath is deprecated and will be replaced by DocTestTextfilePlus.path.
E   see https://docs.pytest.org/en/latest/deprecations.html#node-fspath-in-favor-of-pathlib-and-node-path
```

pllim:
Fix #155 

https://docs.pytest.org/en/latest/deprecations.html#node-fspath-in-favor-of-pathlib-and-node-path

### DIFF
--- a/pytest_doctestplus/plugin.py
+++ b/pytest_doctestplus/plugin.py
@@ -187,21 +187,26 @@ def pytest_configure(config):
         # behavior (which doesn't do whitespace normalization or
         # handling __doctest_skip__) doesn't happen.
         def collect(self):
+            if PYTEST_GE_6_3:
+                path = self.path
+            else:
+                path = Path(self.fspath)
+
             # When running directly from pytest we need to make sure that we
             # don't accidentally import setup.py!
-            if self.fspath.basename == "setup.py":
+            if path.name == "setup.py":
                 return
-            elif self.fspath.basename == "conftest.py":
+            elif path.name == "conftest.py":
                 if PYTEST_GE_6_3:
                     module = self.config.pluginmanager._importconftest(
-                        Path(self.fspath), self.config.getoption("importmode"))
+                        path, self.config.getoption("importmode"))
                 elif PYTEST_GT_5:
                     module = self.config.pluginmanager._importconftest(
                         self.fspath, self.config.getoption("importmode"))
                 else:
                     module = self.config.pluginmanager._importconftest(
                         self.fspath)
-            else:
+            elif not PYTEST_GE_6_3:
                 try:
                     module = self.fspath.pyimport()
                 except ImportError:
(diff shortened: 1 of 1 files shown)


pllim:
lol... sigh... I guess I need to look at the warning more carefully.

pytest_doctestplus/plugin.py
```diff
@@ -187,28 +187,35 @@ class DocTestModulePlus(doctest_plugin.DoctestModule):
...
+            elif not PYTEST_GE_6_3:
                 try:
                     module = self.fspath.pyimport()
                 except ImportError:
                     if self.config.getvalue("doctest_ignore_import_errors"):
                         pytest.skip("unable to import module %r" % self.fspath)
                     else:
                         raise
+            else:
+                raise NotImplementedError("unable to parse %r" % path)
```
pllim, line 218: I don't know how to replace `self.fspath.pyimport()` here, as `pytest` is dropping support for `py.path.local`.

saimn: Hmm, so if `self.path` is a pathlib object I guess we need to import to module with `importlib.import_module` ?

pllim: Yes, I think `self.path` is `pathlib.Path`. Should we ping a pytest dev for review?

saimn: Maybe useful to know if they have a better solution or if we should re-implement the part of `.pyimport` that we need (https://github.com/pytest-dev/py/blob/d4111cae5572809a23d3d5cc88bed8e4af720929/py/_path/local.py#L649-L650)

pllim: @nicoddemus perhaps or at least he'll know who else to ping? 🙏

nicoddemus: Hey!

An alternative is to import the internal [import_path() function we use](https://github.com/pytest-dev/pytest/blob/bfbf733a3ed2f58a71570aa6e950444f3c5b0ea1/src/_pytest/pathlib.py#L456), however this is not ideal in the long term.

That function is a bit long because of the different import modes we support, however if you would like to only support the newer `importlib` mode, the code is relatively small and uses only the `importlib` stable API.

https://github.com/pytest-dev/pytest/blob/bfbf733a3ed2f58a71570aa6e950444f3c5b0ea1/src/_pytest/pathlib.py#L489-L504

cc @RonnyPfannschmidt @asottile in case they have other suggestions.

PR #157, opened 2021-06-13
Update config._getconftest_pathlist for pytest-dev

saimn:
Ref https://github.com/pytest-dev/pytest/pull/7870

Close  #156

### DIFF
--- a/pytest_doctestplus/plugin.py
+++ b/pytest_doctestplus/plugin.py
@@ -14,6 +14,7 @@ from textwrap import indent
 
 import pytest
 from packaging.version import Version
+from _pytest.pathlib import import_path
 
 from pytest_doctestplus.utils import ModuleChecker
 
@@ -210,7 +211,7 @@ def pytest_configure(config):
                         self.fspath)
             else:
                 try:
-                    module = fspath.pyimport()
+                    module = import_path(fspath)
                 except ImportError:
                     if self.config.getvalue("doctest_ignore_import_errors"):
                         pytest.skip("unable to import module %r" % fspath)
--- a/pytest_doctestplus/plugin.py
+++ b/pytest_doctestplus/plugin.py
@@ -189,12 +189,19 @@ def pytest_configure(config):
         def collect(self):
             # When running directly from pytest we need to make sure that we
             # don't accidentally import setup.py!
-            if self.fspath.basename == "setup.py":
+            if PYTEST_GE_6_3:
+                fspath = self.path
+                filepath = self.path.name
+            else:
+                fspath = self.fspath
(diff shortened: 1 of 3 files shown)


pllim:
Something else need changing to be compatible with pytest-dev

```
.../pytest_doctestplus/plugin.py:213: in collect
    module = fspath.pyimport()
E   AttributeError: 'PosixPath' object has no attribute 'pyimport'
```

pytest_doctestplus/plugin.py
```diff
@@ -203,10 +210,10 @@ def collect(self):
                         self.fspath)
             else:
                 try:
-                    module = self.fspath.pyimport()
+                    module = fspath.pyimport()
```
pllim, line 213: This change broke something.

saimn: Yeah, that method doesn't exist on `Path` objects, so I will need to reimplement it I guess.

saimn:
@pllim - Tests pass now ! Also I had forgotten about #156... see the discussion there about `pyimport`, I chose the "import the internal pytest function" solution, seems fine for now.

pytest_doctestplus/plugin.py
```diff
@@ -438,7 +464,14 @@ def pytest_ignore_collect(self, path, config):
             dirpath = Path(path).parent
         else:
             dirpath = path.dirpath()
-        collect_ignore = config._getconftest_pathlist("collect_ignore", path=dirpath)
+
+        try:
+            collect_ignore = config._getconftest_pathlist("collect_ignore", path=dirpath)
+        except TypeError:
+            # Pytest 7.0+
```
pllim, line 471: Why not doing version check here?

pllim:
> Also I had forgotten about #156

Hah, me too! 😸

pytest_doctestplus/plugin.py
saimn: I forgot about it ;). And this is a bit of a mess, some changes are in 6.3 which is not released yet, and some will be in 7.0. And dev is currently 6.3.devXXX though it includes the 7.0 changes apparently.
So I added PYTEST_GT_6_3 after but I'm still not sure if that's a good idea.

saimn: I mean try/except is easy to do, distinguishing 6.3 and 7.0 is currently not obvious.

pllim: Okay, maybe we can merge as is but open issue to revisit in the future because you don't want the stable pytest to keep hitting the `except` block.

pllim (approved):
Thanks!

PR #168, opened 2021-11-15
Compatibility with pytest-dev

saimn:


pytest_doctestplus/plugin.py
```diff
@@ -22,8 +22,9 @@
 
 _pytest_version = Version(pytest.__version__)
 PYTEST_GT_5 = _pytest_version > Version('5.9.9')
+PYTEST_GE_5_4 = _pytest_version >= Version('5.4')
 PYTEST_GE_6_3 = _pytest_version.is_devrelease or _pytest_version >= Version('6.3')
-PYTEST_GT_6_3 = _pytest_version.is_devrelease or _pytest_version > Version('6.3')
+PYTEST_GE_7_0 = _pytest_version.is_devrelease or _pytest_version >= Version('7.0')
```
saimn, line 27: It seems 7.0 will be the next version and there will not be a 6.3 release, but we can cleanup this later.

pllim (approved):
If the CI pass, LGTM. Thanks!

PR #169, opened 2021-11-22
Add comment characters for Markdown

lpsinger:


pllim:
Thanks! But what is the motivation for this patch and can you please add a test?

lpsinger:
The motivation is that I have doctests in my Markdown README file in https://github.com/skyportal/healpix-alchemy, and I am tired of the warning, `file format '.md' is not recognized, assuming '\.\.' as the comment character`.

lpsinger:
I added some tests.

saimn (approved):
Looks good, thanks @lpsinger.

PR #175, opened 2022-02-15
Doctests in ufuncs should respect __doctest_skip__

lpsinger:


mhvk (approved):
Looks good to me! Definitely makes sense to pass it through the same selection criteria

pllim:
I think we should document how to use `__doctest_skip__` to skip unwanted ufunc doctest?

lpsinger:
Following recent discussions, I have added back the `--doctest-ufunc` option, which is disabled by default.

pllim:
Re-running astropy/astropy#12853 🤞

pllim:
So.... do we *not* want to use it over at `astropy` then? If we do, how do we do it without astropy/astropy#12861 ?

lpsinger:
> So.... do we _not_ want to use it over at `astropy` then? If we do, how do we do it without [astropy/astropy#12861](https://github.com/astropy/astropy/pull/12861) ?

You could revert [#12861](https://github.com/astropy/astropy/pull/12861) if you wanted.

pllim:
@mhvk , do you really want ufunc doctesting in astropy ?

mhvk:
We don't have that many ufuncs, and I don't think any have examples that would be tested, since they are really for internal use. So, certainly no hurry. On the other hand, long-term we might as well test things in case things change.

pllim (approved):
`astropy` core seems fine with this patch. Thanks!

Shall we merge?

nstarman (approved):

pllim:
Thanks!

lpsinger:
Thank you! When do you expect to do the next release of pytest-doctestplus?

pllim:
@lpsinger , #177 is blocking the release.

pllim:
I'll push a release soon. Stay tuned!

pllim:
v0.12.0 is released on PyPI.

PR #192, opened 2022-10-10
Remove an obsolete version check from tests

eerovaher:
One of the tests was skipped with `pytest` versions older than 4.3, but the minimum required `pytest` version is 4.6 since 08060276b4f860fce9b49ce0786fc77cd49ccc09


pllim (approved):
LGTM. Thanks!

PR #198, opened 2023-05-10
DOC: Version check is supported in the doctest-requires directive since v0.5

pllim:
Version check is supported in the `doctest-requires` directive since v0.5 (2019-11-15, #78) but I didn't know until I dug into the change log.

ref: https://github.com/astropy/astropy/pull/14797#discussion_r1189980217


saimn:
Thanks @pllim !

PR #222, opened 2023-09-08
Add pyproject.toml tests for some options

WilliamJamieson:
In https://github.com/astropy/astropy/pull/15247#issuecomment-1703297888, it was requested that `pytest-doctestplus` add unit testing to confirm the support of `doctest_nocursedirs` and `doctest_subpackage_requires`. This PR adds the requested tests.

~Note that this PR requires bumping the minimum `pytest` version from 4.6 to 6. While this is a big jump it is inline with the [Scientific Python Style Guide](https://learn.scientific-python.org/development/guides/pytest/#configuring-pytest). In particular PP302 says that projects should use `pytest` 6 or later. Moreover, the method to test `pyproject.toml` files for `pytest` configuration, [`makepyprojecttoml`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.Pytester.makepyprojecttoml), was only added in `pytest 6.0.0`.~

I added skips on the test cases involving the `pyproject.toml`.

### DIFF
--- a/tests/test_doctestplus.py
+++ b/tests/test_doctestplus.py
@@ -3,6 +3,8 @@ import os
 from textwrap import dedent
 import sys
 
+from packaging.version import Version
+
 import pytest
 
 import doctest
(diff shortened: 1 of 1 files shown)


pllim (approved):
Thanks! Though I want Simon to have a chance to review since he originally brought up the concern.

bsipocz (approved):
Now that it doesn't bump the requirements (that would definitely be a blocker), it looks all good to me. Thanks!

pllim:
Thanks, all!

saimn:
Thanks for adding the tests @WilliamJamieson.

PR #224, opened 2023-09-25
Use super for output checker inheritance

matthew-brett:
Remove previous workarounds for Python 2 old-style classes.

### DIFF
--- a/pytest_doctestplus/output_checker.py
+++ b/pytest_doctestplus/output_checker.py
@@ -43,8 +43,6 @@ class OutputChecker(doctest.OutputChecker):
     rtol = 1e-05
     atol = 1e-08
 
-    _original_output_checker = doctest.OutputChecker
-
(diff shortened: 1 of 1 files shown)


pytest_doctestplus/output_checker.py
```diff
@@ -281,18 +279,13 @@ def check_output(self, want, got, flags):
         if flags & FLOAT_CMP:
             return self.normalize_floats(want, got, flags)
 
-        # Can't use super here because doctest.OutputChecker is not a
-        # new-style class.
-        return self._original_output_checker.check_output(
-            self, want, got, flags)
+        return self(super).check_output(want, got, flags)
```
pllim, line 282: Maybe this?
```suggestion
        return super().check_output(want, got, flags)
```

pllim (approved):
The diff LGTM, provided CI agrees. Thanks!

PR #225, opened 2023-09-26
Precautionary init for OutputChecker

matthew-brett:
CPython source has no __init__ for this class, but just in case, and to
reassure readers.

https://github.com/python/cpython/blob/main/Lib/doctest.py#L1609

### DIFF
--- a/pytest_doctestplus/output_checker.py
+++ b/pytest_doctestplus/output_checker.py
@@ -53,9 +53,6 @@ class OutputChecker(doctest.OutputChecker):
         r"([0-9]+)L", re.UNICODE)
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


pllim (commented):
I don't know how I feel about inserting unnecessary constructor call.

Do they have plans to add init after so long without? Surely that would be announced in change log?

matthew-brett:
I could replace with a comment to remember to check the CPython change log in case they introduce an `__init__`?

bsipocz (commented):
I would say go ahead with this one, if it fails as they add one, we'll notice when we add a new python version  support. 

However, if they add one, and we don't have this, we won't notice automatically (and we may want to use that init).

pllim (approved):
Okay, then. Thanks everyone!

PR #228, opened 2023-10-06
BUG: Fix module level import requires

seberg:
Since the logic was written in a negate way compared to the normal skipping, the module level skip didn't take into account the special "*" and "." patterns.

gh-109 probably misdiagnosed the module actually missing only on the run the had pytest-xdist: the test does not seem to ever pass either way.

Closes gh-109

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 1.1.0 (unreleased)
 ==================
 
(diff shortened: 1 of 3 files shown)


bsipocz (approved):
Thanks for fixing this long-standing bug. 

This would ideally need a changelog entry, but I'll just go ahead and push that to the PR and merge.

PR #248, opened 2024-04-09
TST: add tests for ufuncs without docstrings, wrapped or not

lpsinger:
This is known to fail on Python 3.11.9 because `--doctest-ufunc` was broken for wrapped ufuncs without docstrings by https://github.com/python/cpython/pull/115440.


bsipocz (approved):
Thanks!

lpsinger:
Reported to CPython: https://github.com/python/cpython/issues/117692

PR #249, opened 2024-04-16
Update skips for broken test

lpsinger:


tests/test_doctestplus.py
```diff
@@ -1142,7 +1142,7 @@ def f():
 
 
 @pytest.mark.xfail(
-        python_version() == '3.11.9',
+        python_version() in {'3.11.9', '3.12.3'},
```
pllim, line 1145: Unlikely we will introduce duplicate entries here, so `set` isn't strictly necessary?

```suggestion
        python_version() in ('3.11.9', '3.12.3'),
```

Thanks!

pllim (approved):
Thanks!

PR #253, opened 2024-06-27
BUG: Fix multiline code in generate/update diff

seberg:
The code may have more than one line, so need to account for that.

Closes gh-252


bsipocz (approved):
Thank you so much for this fix!

I'll push a commit for the changelog and then merge.

PR #254, opened 2024-06-28
MAINT: Remove accidental debug print from new tests

seberg:
I forgot to delete these lines (the pytest print removes the interesting stuff, so it's hard to see what differs just from them).


pllim (approved):
Thanks!

PR #310, opened 2025-11-01
TST: Debug and fix new Windows + Python 3.14 failure (try 2)

pllim:
Works around https://github.com/scientific-python/pytest-doctestplus/issues/305 , reverts https://github.com/scientific-python/pytest-doctestplus/pull/307

Was https://github.com/scientific-python/pytest-doctestplus/pull/308

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,14 @@
+import sys
+from packaging.version import Version
+from platform import python_version
+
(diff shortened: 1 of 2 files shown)


pllim:
I guess it does have to be at the top level. Weird!

Thanks!

PR #322, opened 2026-03-17
TST: skip test more roboustly for python version

bsipocz:
New bugfix release is out and thus the cron is failing, so I just fixed up the test a bit better so we don't need to keep adding to the list.


pllim (approved):
Thanks!

